PR #27, opened 2022-07-18
Solution

4ndyua:


cinema/views.py
```diff
@@ -76,16 +86,18 @@ def get_serializer_class(self):
         return MovieSerializer
 
 
-class MovieSessionViewSet(viewsets.ModelViewSet):
+class MovieSessionViewSet(mixins.ListModelMixin, mixins.CreateModelMixin, mixins.RetrieveModelMixin,
```
KatiukhaO, line 89: It`s correct. But, I think, you should to stat this view Model View Set, because, you don't change behavior when you use all generics mixins

cinema_service/urls.py
```diff
@@ -5,4 +5,5 @@
     path("admin/", admin.site.urls),
     path("api/cinema/", include("cinema.urls", namespace="cinema")),
     path("__debug__/", include("debug_toolbar.urls")),
+    path("api/user/", include("user.urls", namespace="user")),
```
KatiukhaO, line 8: I'm not sure, but how to set the order of endpoints so that debug is last? I just read the code better when the debug is over

KatiukhaO (approved):

MasterpieceElbow (approved):

PR #30, opened 2022-07-19
Solution

vasylhnatiuk:


cinema/views.py
```diff
@@ -2,9 +2,16 @@
 
 from django.db.models import F, Count
 from rest_framework import viewsets
+from rest_framework.authentication import TokenAuthentication
+from rest_framework.mixins import ListModelMixin, \
```
KatiukhaO, line 6: When you have a long tape of imports, it is better to put all these imports in brackets and make a transfer

cinema/views.py
```diff
@@ -76,7 +96,14 @@ def get_serializer_class(self):
         return MovieSerializer
 
 
-class MovieSessionViewSet(viewsets.ModelViewSet):
+class MovieSessionViewSet(
```
KatiukhaO, line 99: It`s correct. But, I think, you should to stat this view Model View Set, because, you don't change behavior when you use all generics mixins

KatiukhaO (approved):

MasterpieceElbow (approved):

PR #32, opened 2022-07-19
'Solution'

Yurnerroo:


cinema/views.py
```diff
@@ -117,12 +148,18 @@ class OrderPagination(PageNumberPagination):
...
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Order.objects.all().prefetch_related(
         "tickets__movie_session__movie", "tickets__movie_session__cinema_hall"
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
daria-niklaevna, line 162: We have task №4: We should give the ability for authenticated users to create order

MasterpieceElbow (approved):

PR #67, opened 2022-11-21
Solution

olia-trofymchuk:


cinema/views.py
```diff
@@ -21,24 +24,45 @@
...
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.ListModelMixin, mixins.CreateModelMixin, viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = [IsAdminOrIfAuthenticatedReadOnly]
```
mltdll, line 33: For some reason, you use a list here, but tuples for other settings like this.

mltdll (approved):

cinema/views.py
Nattalli: I wouldn't have even noticed it, but I agree, it's strange 😄

Nattalli: It's not a mistake, but try to stick to one code style in the future 🤝

Nattalli (approved):
👍🏻

PR #74, opened 2022-11-22
Solution

linaSam:


cinema/views.py
```diff
@@ -118,12 +148,23 @@ class OrderPagination(PageNumberPagination):
...
+):
     queryset = Order.objects.all().prefetch_related(
         "tickets__movie_session__movie", "tickets__movie_session__cinema_hall"
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def get_permissions(self):
```
wQuelS, line 164: i don't know if it is a mistake, but this class already inherits from CreateModelMixin and has his own permission class. For me this method seems a bit redundant. If you don't agree just skip my reject, and let's wait what mento says :)

wQuelS (changes requested):
I really like your code. I've left only one comment

cinema/views.py
linaSam: agree - seems that it doesn't do much here) thanks!

linaSam: i agree with you - seems like it doesn't do much here) thanks!

Nattalli (approved):
good job!

PR #86, opened 2022-11-22
Solution!

ketstap162:


wQuelS (changes requested):
I guess you have to remove json file with data from your PR. Also, i guess , it is not a mistake, but it is better to use tuples instead of lists, if you can

wQuelS (approved):

cinema/models.py
```diff
@@ -94,7 +94,8 @@ def validate_ticket(row, seat, cinema_hall, error_to_raise):
             if not (1 <= ticket_attr_value <= count_attrs):
                 raise error_to_raise(
                     {
-                        ticket_attr_name: f"{ticket_attr_name} number must be in available range: "
+                        ticket_attr_name: f"{ticket_attr_name} "
+                        f"number must be in available range: "
```
Y-Havryliv, line 98: don't overuse f-strings

user/serializers.py
```diff
@@ -1 +1,29 @@
...
+        return get_user_model().objects.create_user(**validated_data)
+
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password", None)
+        user = super().update(instance, validated_data)
+
+        if password:
+            user.set_password(password)
+
+        user.save()
```
Y-Havryliv, line 27: it's better to save user in the if-statement

Y-Havryliv (changes requested):

user/serializers.py
ketstap162: I think it's bad idea, because we will not be able to change other user data without changing his password.

Nattalli: but according to your code: if there is no password, then the user is updated anyway... It seems not OK

Nattalli (changes requested):

user/serializers.py
ketstap162: `if password: user.set_password`
It's not password check. It's just update... You can write other password, not valid, and user password will be update

Nattalli (approved):

PR #102, opened 2022-12-10
Solution

Vlad-Peresta:


user/permissions.py
```diff
@@ -0,0 +1,13 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
+        )
```
Serhii-Leonenko, line 13: permissions should be in cinema/

Serhii-Leonenko (changes requested):

user/permissions.py
Vlad-Peresta: Ok, but I thought that permissions related with user should be in /user

OleksandrPozden (approved):

PR #118, opened 2023-02-01
Solution

ansicat:


cinema/views.py
```diff
@@ -24,21 +27,61 @@
...
+    authentication_classes = (TokenAuthentication,)
+
+    def get_permissions(self):
+        if self.action in ("list", "create"):
+            return [
+                permission()
+                for permission in [IsAdminOrIfAuthenticatedReadOnly]
+            ]
+
+        raise exceptions.NotFound
```
ArsenPidhoretskyi: This is not needed. You need to change inheritance
You can use mixins: https://www.django-rest-framework.org/api-guide/generic-views/#mixins
And then add permission_classes = (IsAdminOrAuthenticatedOrReadOnly, )
**Fix in all places**

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #119, opened 2023-02-08
Solution

Nexxel05:


cinema/views.py
```diff
@@ -118,12 +149,18 @@ class OrderPagination(PageNumberPagination):
...
+    mixins.CreateModelMixin,
+    mixins.ListModelMixin,
+    GenericViewSet
+):
     queryset = Order.objects.prefetch_related(
         "tickets__movie_session__movie", "tickets__movie_session__cinema_hall"
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication, )
```
Bohdan-Salamakha: add it todefault django authentification methods to evade duplicates

user/serializers.py
```diff
@@ -1 +1,28 @@
...
+        read_only_fields = ("id", "is_staff")
+        extra_kwargs = {
+            "password": {
+                "write_only": True,
+                "min_length": 5,
+            }
+        }
+
+    def create(self, validated_data):
+        return get_user_model().objects.create_user(**validated_data)
```
Bohdan-Salamakha: duplicated get_user_model()
add it before meta class like user_class = get_user_model()

Bohdan-Salamakha (changes requested):
v1

user/serializers.py
```diff
@@ -1 +1,30 @@
-# write your code here
+from django.contrib.auth import get_user_model
+from rest_framework import serializers
+
+
+class UserSerializer(serializers.ModelSerializer):
+    user_class = get_user_model()
+
+    class Meta:
+        model = get_user_model()
```
Bohdan-Salamakha, line 9: why duplicated again?...
```
class UserSerializer(serializers.ModelSerializer):
    user_class = get_user_model()

    class Meta:
        model = UserSerializer.user_class 
```

Bohdan-Salamakha (changes requested):
v2

user/serializers.py
Nexxel05: I don't get it. I can not refer neither to user_class or UserSerializer.user_class within class Meta.

Bohdan-Salamakha: yeah agreed, it will be harder to call it from inner class, but it's possible
approved

Bohdan-Salamakha (approved):

PR #130, opened 2023-03-07
Solution

Paul-Starodub:


cinema/models.py
```diff
@@ -12,52 +14,56 @@ class CinemaHall(models.Model):
...
-    def __str__(self):
-        return self.first_name + " " + self.last_name
+    def __str__(self) -> str:
+        return self.full_name
 
     @property
-    def full_name(self):
-        return f"{self.first_name} {self.last_name}"
+    def full_name(self) -> str:
+        return self.first_name + " " + self.last_name
```
ArsenPidhoretskyi: No no return to f-string
It is more better that concatenate

ArsenPidhoretskyi (changes requested):

cinema/models.py
Paul-Starodub: Ok

ArsenPidhoretskyi (approved):
well done!

PR #139, opened 2023-03-22
Solution

derkach55:


ArsenPidhoretskyi (approved):
Good job. But typing does not exist

PR #149, opened 2023-04-05
Develop

Zoriana-Dvulit:


user/views.py
```diff
@@ -1 +1,25 @@
...
+    permission_classes = [AllowAny, ]
+
+
+class CreateTokenView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+
+
+class ManageUserView(generics.RetrieveUpdateAPIView):
+    serializer_class = UserSerializer
+    authentication_classes = (TokenAuthentication, )
```
olena-hul: I think it will work without specifying authentication_classes explicitly, because you have already defined TokenAuthentication as the default authentication class in settings.py

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #156, opened 2023-04-08
Solution

SergiiMachulin:


Morty67 (approved):

user/permissions.py
```diff
@@ -0,0 +1,13 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view) -> bool:
+        return bool(
+            (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
+        )
```
YuriiKindrat: i think it is better to move this file to the cinema app, cause u are using this permission only there

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #165, opened 2023-04-10
Solution

inrock1:


kuznetsb (approved):
LGTM

cinema/permissions.py
```diff
@@ -0,0 +1,17 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
```
wQuelS, line 4: Don't you wanna use it in your views ?

wQuelS (changes requested):

cinema/permissions.py
inrock1: I use it in views, just do it through the definitions of the default permission.
image

Serhii-Leonenko (approved):

PR #166, opened 2023-04-10
solution

amber-marichi:


inrock1 (approved):
Nice.

cinema/views.py
```diff
@@ -118,12 +147,23 @@ class OrderPagination(PageNumberPagination):
...
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication, )
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
+
+    def get_permissions(self):
+        if self.action == "create":
+            return [IsAuthenticated()]
+        return super().get_permissions()
```
Serhii-Leonenko: For what purpose do you need this method here, I think IsAuthenticated is ok for this case if you use it like permission_class attribute

Serhii-Leonenko (changes requested):

cinema/views.py
amber-marichi: you are right, seems i was trying too hard to combine IsAuthenticated with custom permission class :) fixed now

wQuelS (approved):
Good

PR #177, opened 2023-04-10
Learn flow

NEkropo1:


ada-krav (approved):

cinema/views.py
```diff
@@ -131,8 +147,14 @@ def get_queryset(self):
...
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    def get_permissions(self):
+        if self.action == "create":
+            self.permission_classes = (IsAuthenticated,)
+        else:
+            self.permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+        return [permission() for permission in self.permission_classes]
```
YuriiKindrat: why do u need this? u can use IsAuthenticated for all methods here, cause list is in SAFE_METHODS

cinema_service/permissions.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+SAFE_LIST = (
+    "list",
+    "create",
+)
```
YuriiKindrat: why do u need this ?

YuriiKindrat (changes requested):

cinema_service/permissions.py
NEkropo1: Actually not sure how it worked, but I remember that something wasn't explicitly managed in default safe_methods, so I redefined them, I'll fix it in a while

NEkropo1: Yup that was just testing on how can we redefine permissions, totally redundant )

cinema/views.py
NEkropo1: Actually I was thinking about other methods, pretty long time, till I mentioned parent classes)
Should be more scrupulous, for sure.

Serhii-Leonenko (approved):

PR #178, opened 2023-04-10
Solution

oksana-feshchenko:


cinema/views.py
```diff
@@ -21,24 +24,49 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
```
ZhAlexR, line 31: you can just use `generics.ListCreateAPIView` here instead all classes which were inherited

cinema/views.py
```diff
@@ -118,12 +148,18 @@ class OrderPagination(PageNumberPagination):
...
-class OrderViewSet(viewsets.ModelViewSet):
+class OrderViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet,
+):
     queryset = Order.objects.prefetch_related(
         "tickets__movie_session__movie", "tickets__movie_session__cinema_hall"
     )
     serializer_class = OrderSerializer
```
ZhAlexR, line 159: you do need to provide `serializer_class` here if you will overload this in `get_serializer_class` below.

ZhAlexR (approved):

cinema/views.py
Serhii-Leonenko: but after that, you cannot use routers

Serhii-Leonenko (approved):

PR #180, opened 2023-04-10
Develop

ostapT:


danylo-d (approved):

cinema/views.py
```diff
@@ -76,7 +104,14 @@ def get_serializer_class(self):
...
 
 
-class MovieSessionViewSet(viewsets.ModelViewSet):
+class MovieSessionViewSet(
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.CreateModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+    viewsets.GenericViewSet,
```
YuriiKindrat: no need to change inheritance here

cinema/views.py
```diff
@@ -87,6 +122,14 @@ class MovieSessionViewSet(viewsets.ModelViewSet):
...
     )
     serializer_class = MovieSessionSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+
+    def get_permissions(self):
+        if self.action in ("create", "update", "partial_update", "destroy"):
+            return [IsAdminUser()]
+
+        return super().get_permissions()
```
YuriiKindrat: why do u need this ?

YuriiKindrat (changes requested):

cinema/views.py
ostapT: for some reason I thought that it was necessary to override the permissions for specific actions here separately. But I already understood that we already define permission_classes = (IsAdminOrIfAuthenticatedReadOnly,) for this class

YuriiKindrat (approved):

PR #205, opened 2023-05-09
Solution

avkpol:


ArsenPidhoretskyi (changes requested):
Tests are not passed
image
Also this is not completed

avkpol:
code is corrected, permissions are customized, and what else? Why do codes pass in Pycharm and fail in git? Are tests the same? How can I check in Pycharm if I make all correct?

ArsenPidhoretskyi:
Ca nyou pin a screenshot with tests?
Now you does not complete the task. See description. You must allow only this apis. You must change ModelView set to another View

ArsenPidhoretskyi (changes requested):
Changes requested

vasylhnatiuk (approved):
Good job

vasylhnatiuk (approved):

PR #207, opened 2023-05-11
Develop

alina-boichenko:


cinema/views.py
```diff
@@ -136,3 +157,8 @@ def get_serializer_class(self):
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    def get_permissions(self):
+        if self.action not in ("list", "create"):
+            raise NotFound
```
ArsenPidhoretskyi: It is not true. It is mean you can not find object by id. Which is not true in this case
image

ArsenPidhoretskyi (changes requested):

user/serializers.py
```diff
@@ -1 +1,24 @@
...
+        extra_kwargs = {"password": {"write_only": True, "min_length": 5}}
+
+    def create(self, validated_data):
+        return get_user_model().objects.create_user(**validated_data)
+
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password", None)
+        user = super().update(instance, validated_data)
+
+        if password:
```
Grayder0152, line 20:
```suggestion
        if password is not None:
```

Grayder0152 (approved):

PR #208, opened 2023-05-13
Solution

Paul-Maslov:


Bohdan-Salamakha (changes requested):
fix tests

cinema/views.py
```diff
@@ -21,24 +26,32 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(generics.ListCreateAPIView):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
Bohdan-Salamakha: add TokenAuthentication to the default authentication_classes in settings.py

Bohdan-Salamakha (changes requested):
v1

cinema/views.py
Paul-Maslov: But if i do it, the original 3 tests will fall. They raise errors but with another code

Bohdan-Salamakha (changes requested):
fix flake8

Bohdan-Salamakha (approved):

PR #216, opened 2023-06-01
Solution

andrii-skryl:


cinema/views.py
```diff
@@ -75,6 +87,11 @@ def get_serializer_class(self):
 
         return MovieSerializer
 
+    def get_permissions(self):
+        if self.action not in ("list", "create", "retrieve"):
```
olena-hul: You can just use List, Create, Retrieve mixins

cinema/views.py
```diff
@@ -136,3 +157,8 @@ def get_serializer_class(self):
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    def get_permissions(self):
+        if self.action not in ("list", "create"):
```
olena-hul: Use List and Create mixins

olena-hul (changes requested):

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #239, opened 2023-07-09
Solution

andriy-md:


vsmutok (changes requested):

cinema/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+    def has_permission(self, request, view):
+        if (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+        ):
+            return True
+        if request.user and request.user.is_staff:
+            return True
+        return False
```
ArsenPidhoretskyi: Write this by usage or statement

ArsenPidhoretskyi: And also use only return

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #249, opened 2023-07-21
solution

mblazhko:


cinema/views.py
```diff
@@ -88,6 +101,11 @@ class MovieSessionViewSet(viewsets.ModelViewSet):
     )
     serializer_class = MovieSessionSerializer
 
+    def get_permissions(self):
+        if self.action not in ("list", "create", "retrieve"):
+            return [IsAdminUser()]
+        return super().get_permissions()
```
olena-hul: just include the needed mixins, no need to override get_permissions

olena-hul (changes requested):

cinema/views.py
olena-hul: You should make available only 3 actions, list, create and retrieve

cinema/views.py
```diff
@@ -118,12 +131,17 @@ class OrderPagination(PageNumberPagination):
     max_page_size = 100
 
 
-class OrderViewSet(viewsets.ModelViewSet):
```
olena-hul, line 121: You should make available only 3 actions for MovieSessionViewSet: list, create and retrieve

olena-hul (changes requested):

cinema/views.py
mblazhko: image
nope, i need all actions and ModelViewSet inherits all i need

> You should make available only 3 actions for MovieSessionViewSet: list, create and retrieve

Bohdan-Salamakha (approved):

PR #251, opened 2023-07-24
Solution

YehorKorn:


cinema/views.py
```diff
@@ -124,15 +171,28 @@ class OrderViewSet(viewsets.ModelViewSet):
...
         return Order.objects.filter(user=self.request.user)
 
     def get_serializer_class(self):
         if self.action == "list":
             return OrderListSerializer
+        if self.action == "create":
+            return OrderSerializer
 
-        return OrderSerializer
+        return Response()
```
ArsenPidhoretskyi: It is always must return Serializer

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

cinema/views.py
```diff
@@ -24,21 +28,62 @@
...
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
+
+    def update(self, request, *args, **kwargs):
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+    def destroy(self, request, *args, **kwargs):
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+    def retrieve(self, request, *args, **kwargs):
+        return Response(status=status.HTTP_404_NOT_FOUND)
```
ArsenPidhoretskyi: Why do you need to hardcode just change inheritance. Fix in all places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #253, opened 2023-07-30
Solution

diana-shyrokikh:


cinema/views.py
```diff
@@ -120,10 +144,13 @@ class OrderPagination(PageNumberPagination):
 
 class OrderViewSet(viewsets.ModelViewSet):
```
YuriiKindrat: does orderview set could only create&list ?
image

YuriiKindrat (changes requested):
it seems to me, that it will be much easier to use generics to remove some methods from available in view, but it is up to u ))

Oleg-s-s (approved):

PR #258, opened 2023-08-01
'Solution'

ant-komarov:


cinema/views.py
```diff
@@ -118,12 +148,23 @@ class OrderPagination(PageNumberPagination):
...
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+
+    def get_permissions(self):
+        if self.action == "create":
+            return [IsAuthenticated()]
+        return super().get_permissions()
```
YuriiKindrat: it seems to me, that u can use IsAuthenticated permission for all actions avaliable on this view )

YuriiKindrat (changes requested):

Oleg-s-s (approved):

PR #260, opened 2023-08-02
'Solution'

yakhlamova:


cinema/views.py
```diff
@@ -136,3 +179,9 @@ def get_serializer_class(self):
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    def get_permissions(self):
+        if self.action == "create":
+            return [IsAuthenticated()]
+
+        return super().get_permissions()
```
YuriiKindrat: it seems to me, that u can just use isAuthenticated for all actions available on this view )

YuriiKindrat (changes requested):

Oleg-s-s (approved):

PR #274, opened 2023-08-03
Solution

Tarasidze:


cinema/views.py
```diff
@@ -21,24 +31,49 @@
...
+):
     queryset = CinemaHall.objects.all()
     serializer_class = CinemaHallSerializer
+    authentication_classes = (TokenAuthentication, )
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
 
 
-class MovieViewSet(viewsets.ModelViewSet):
+class MovieViewSet(
+    viewsets.GenericViewSet,
```
LiudmylaKulinchenko: you should inherit from viewsets.GenericViewSet at the end of the list of base classes to ensure proper method resolution and consistent behavior

LiudmylaKulinchenko: mind everywhere

LiudmylaKulinchenko (changes requested):

cinema/views.py
Tarasidze: really!!
i forgot that mixins should be in the first place of inheritance))

LiudmylaKulinchenko (approved):
Good job 💪😎

PR #278, opened 2023-08-04
Solution

bezhevets:


cinema/views.py
```diff
@@ -118,7 +137,11 @@ class OrderPagination(PageNumberPagination):
     max_page_size = 100
 
 
-class OrderViewSet(viewsets.ModelViewSet):
+class OrderViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet,
+):
     queryset = Order.objects.prefetch_related(
```
ArsenPidhoretskyi, line 144: Add permissions due to 
OrderViewSet - We should give the ability for authenticated users to create order

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #292, opened 2023-08-22
Solution

arturiermolenko:


Bohdan-Salamakha (changes requested):
add IsAdminOrIfAuthenticatedReadOnly to the default permission_classes in settings.py

Bohdan-Salamakha (approved):

PR #293, opened 2023-08-24
Solution

Oleksiy-Lyashenko:


user/serializers.py
```diff
@@ -1 +1,32 @@
...
+        extra_kwargs = {"password": {"write_only": True, "min_length": 5}}
+
+    def create(self, validated_data):
+        user = User(
+            email=validated_data["email"],
+            username=validated_data["username"]
+        )
+        user.set_password(validated_data["password"])
+        user.save()
+        return user
```
pavlejviki: you can use create_user method() to simplify the create method,  that handles the creation of user instances correctly, including password hashing.

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #294, opened 2023-08-29
Solution

Psheiuk-Nazar:


Bohdan-Salamakha (changes requested):
add IsAdminOrIfAuthenticatedReadOnly to the default permission_classes in settings.py

Polyakiv-Andrey (approved):
Good job!

PR #299, opened 2023-09-07
Develop

TarasSavchyn:


user/permissions.py
```diff
@@ -0,0 +1,13 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if request.user and request.user.is_staff:
+            return True
+
+        if request.user and request.user.is_authenticated:
+            if request.method == "GET":
+                return True
+
+        return False
```
ArsenPidhoretskyi: Use inline return

user/permissions.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework import permissions
+
+
+class IsAdminOrIfAuthenticatedReadOnly(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if request.user and request.user.is_staff:
+            return True
+
+        if request.user and request.user.is_authenticated:
+            if request.method == "GET":
```
ArsenPidhoretskyi: You can use safe methods

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #311, opened 2023-10-11
Develop

eduardhabryd:


OlegatorLE (approved):
Good Job, but you use permission.py only in cinema app. Why is this file locate in user app?

eduardhabryd:
> Good Job, but you use permission.py only in cinema app. Why is this file locate in user app?

I though it makes sense to place pemissions in the user app because it's related to users and our other apps can also use those permissions.

Serhii-Leonenko (approved):

PR #321, opened 2023-10-14
Solution

CaCtUs1k:


Illia-Kononenko (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,42 @@
...
+            elif request.user.is_staff:
+                return True
+        return False
+
+
+class IfAuthReadCreateOnly(BasePermission):
+    def has_permission(self, request, view):
+        if view.action in ["list", "create"] and request.user.is_authenticated:
+            return True
+        return False
```
Serhii-Leonenko: these all permissions are redundant, you could use only one for such purpose

Serhii-Leonenko (changes requested):

cinema/permissions.py
CaCtUs1k: idk why but if i copy permission's code and paste directly in views, checks failed

Serhii-Leonenko (changes requested):
create only one permission class 

image

you don't need all these permissions

cinema/permissions.py
Serhii-Leonenko: create only one permission class 

image

you don't need all these permissions

CaCtUs1k: sorry, my bad!

Sofbog (approved):
LGTM

PR #338, opened 2023-10-18
Solution

OleksandrMoroz24:


cinema/views.py
```diff
@@ -87,6 +115,8 @@ class MovieSessionViewSet(viewsets.ModelViewSet):
         )
```
Sofbog, line 115: do  ` MovieSessionViewSet` needs mixins?

user/views.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from rest_framework import generics
+from rest_framework.authentication import TokenAuthentication
+from rest_framework.permissions import IsAuthenticated
+
+from user.serializers import UserSerializer
+
```
Sofbog, line 8: how can we get TOKEN

Sofbog (changes requested):

Serhii-Leonenko (approved):

PR #339, opened 2023-10-19
Solution py-token-authentication

Mazur-Mariia-Python:


YukihiroSM (changes requested):
Comments in chat

cinema/views.py
```diff
@@ -21,24 +32,54 @@
...
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = (TokenAuthentication, )
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
+
+    def get_object(self):
+        obj = get_object_or_404(self.get_queryset(), pk=self.kwargs["pk"])
+        self.check_object_permissions(self.request, obj)
+        return obj
```
EdAlekseiev: Why do you need this method?

cinema/views.py
```diff
@@ -118,21 +161,31 @@ class OrderPagination(PageNumberPagination):
...
-
         return OrderSerializer
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    def get_permissions(self):
+        if self.action in ("create", ):
+            return [IsAuthenticated()]
+        return super().get_permissions()
```
EdAlekseiev: Why do you need this code method?

EdAlekseiev (changes requested):

cinema/views.py
Mazur-Mariia-Python: Because I should  give the ability for authenticated users to create order. 
And if I remove this method I get following error: 
Failure
Traceback (most recent call last):
  File "C:\Users\Expert\PycharmProjects\py-token-authentication\cinema\tests\test_order_api.py", line 65, in test_post_order
    self.assertNotEqual(response.status_code, status.HTTP_403_FORBIDDEN)
AssertionError: 403 == 403

cinema/views.py
Mazur-Mariia-Python: I'm sorry I forgot to remove this method.

cinema/views.py
ArsenPidhoretskyi: In this case you can just change 
permission_classes = (IsAdminOrIfAuthenticatedReadOnly, ) to 
permission_classes = (IsAuthenticated, )

cinema/views.py
ArsenPidhoretskyi: Still exists

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #348, opened 2023-11-07
Token Authentication API Solution

o-davydova:


cinema/views.py
```diff
@@ -21,24 +23,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
pavlejviki: to avoid code duplication add TokenAuthentication to the default authentication_classes in settings.py and add IsAdminOrIfAuthenticatedReadOnly to the default permission_classes in settings.py

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #358, opened 2023-11-27
Solution

VasylTurok:


cinema/views.py
```diff
@@ -21,24 +24,49 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.CreateModelMixin,
+    mixins.ListModelMixin
```
Dimosphen1, line 30: As a general convention, mixins should be placed before core classes. Take it into account in the future.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #366, opened 2023-12-04
solution

kostomeister:


oleksandr-pupchenko (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import permissions
+
+
+class IsAdminOrIfAuthenticatedReadOnly(permissions.BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            request.method in
+            permissions.SAFE_METHODS and request.user.is_authenticated
```
NEkropo1: `getattr(request.user, 'is_authenticated', False)`
I'm not sure it would work correctly if we won't be authenticated, it always should work as expected, and we don't probably expect AttributeError.

user/serializers.py
```diff
@@ -1 +1,24 @@
...
+
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password")
+        user = super().update(instance, validated_data)
+
+        if password:
+            user.set_password(password)
+            user.save()
+
+        return user
```
NEkropo1, line 26: acid

NEkropo1 (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🔥

PR #367, opened 2023-12-04
py-token-authentication Solution

RVChornyy:


SevKrok (approved):
Nice 👍

user/serializers.py
```diff
@@ -1 +1,22 @@
 # write your code here
```
Y-Havryliv: this is a bad comment for the final code, delete it

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #370, opened 2023-12-04
Solution

Kateryna-Bordonos17:


katyakoziar (approved):
Good job!

user/serializers.py
```diff
@@ -1 +1,23 @@
...
+
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password", None)
+        user = super().update(instance, validated_data)
+
+        if password:
+            user.set_password(password)
+            user.save()
+
+        return user
```
NEkropo1: Acid. If we update user info, but won't update pass, it shouldn't update info.

NEkropo1 (changes requested):

user/serializers.py
```diff
@@ -1 +1,26 @@
...
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password", None)
+        user = get_user_model().objects.select_for_update().get(pk=instance.pk)
+        for attr, value in validated_data.items():
+            setattr(user, attr, value)
+        user.save()
+
+        if password:
+            user.set_password(password)
+            user.save()
```
LiudmylaKulinchenko: U can use save once outside on the condition

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 💪

PR #373, opened 2023-12-04
solution

Naz-iv:


vshvanska (approved):

user/parmitions.py
Y-Havryliv, line 1: it's better to name it `permissions`)

Y-Havryliv: the code is good

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #376, opened 2023-12-04
Solution

Daniil-Pankieiev:


RVChornyy (approved):

user/urls.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from django.urls import path
+
+from user.views import CreateUserView, CreateTokenView, ManageUserView
+
+urlpatterns = [
+    path("register", CreateUserView.as_view(), name="create"),
```
Y-Havryliv: your code is not one-styled, it's better to use `/` at the end of endpoints

Y-Havryliv (changes requested):

user/urls.py
Daniil-Pankieiev: My mistake, fixed

Y-Havryliv (approved):

PR #377, opened 2023-12-04
'Solution'

fedorov-s-od:


DenPrislipskyi (approved):

cinema/views.py
```diff
@@ -21,24 +24,62 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
```
Y-Havryliv, line 27: it's better to move it to a separate module.
What if you need 10 more permissions? Your views would be enormous

Y-Havryliv (changes requested):

PR #382, opened 2023-12-05
Solution

MiskoRuslan:


user/urls.py
```diff
@@ -1 +1,13 @@
-# write your code here
+from django.urls import path
+
+from rest_framework.authtoken import views
```
Y-Havryliv: you don't need this line of code

Y-Havryliv (changes requested):

user/urls.py
MiskoRuslan: done 👍

Y-Havryliv (approved):

PR #384, opened 2023-12-05
Solution

Wonsky1:


cinema/permissions.py
```diff
@@ -0,0 +1,27 @@
...
+                and request.user.is_authenticated
+            )
+            or (
+                request.user and request.user.is_staff
+            )
+
+        )
+
+
+class PermissionMixin:
```
Y-Havryliv: it'd be better to move this mixin into another module(mixins.py or smth like that), and not to mix up mixins and permission

cinema/permissions.py
```diff
@@ -0,0 +1,27 @@
...
+    permission_classes = [IsAdminOrIfAuthenticatedReadOnly]
```
Y-Havryliv: btw, you don't use this mixin, so why do you need it?)

user/serializers.py
```diff
@@ -1 +1,29 @@
...
+
+class UserSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = get_user_model()
+        fields = ("id", "username", "email", "password", "is_staff")
+        read_only_fields = ("id", "is_staff")
+        extra_kwargs = {
+            "password": {
+                "write_only": True,
+                "min_length": 8
```
Y-Havryliv: DUe to the task description password length must be >= 5 symbols

Y-Havryliv (changes requested):

LiudmylaKulinchenko (approved):
Nice work ☃

PR #386, opened 2023-12-06
Solution

OlehShumov:


Oxbay (approved):
Nice)

user/views.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from rest_framework import generics
+from .serializers import UserSerializer
```
Y-Havryliv: import are not ordered correctly.
Imports should be grouped in the following order:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

Y-Havryliv (changes requested):

LiudmylaKulinchenko (approved):
Nice job ☃

PR #412, opened 2024-02-06
Solution

IvanMozhar:


cinema/views.py
```diff
@@ -1,10 +1,20 @@
...
+
+from cinema.models import (
+    Genre,
+    Actor,
+    CinemaHall,
+    Movie,
+    MovieSession,
+    Order
+)
+from cinema.permissions import (IsAdminOrIfAuthenticatedReadOnly, )
```
Dimosphen1, line 17: There is no need to import a single class inside of `()`, you could import it directly. Follow this import approach in the future.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #433, opened 2024-03-05
Solution

Marinel444:


cinema/mixins.py
Dimosphen1, line 1: There is no need to create `PermissionMixin`, it's responsibility lays on the permission class. Remove it

cinema/permissions.py
```diff
@@ -0,0 +1,26 @@
+from django.http import Http404
+from rest_framework.permissions import BasePermission
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+
+    def has_permission(self, request, view):
+        return bool(
+            (
+                view.action in ["retrieve", "list"]
```
Dimosphen1: Reformat this line to check request instead of view
```suggestion
                request.method in SAFE_METHODS
```

cinema/permissions.py
```diff
@@ -0,0 +1,26 @@
...
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
+        )
+
+
+class OrderPermission(BasePermission):
```
Dimosphen1: There is no need for this permission class, having `IsAdminOrIfAuthenticatedReadOnly` is sufficient

cinema/views.py
```diff
@@ -2,9 +2,17 @@
 
 from django.db.models import F, Count
 from rest_framework import viewsets
+from rest_framework.authentication import TokenAuthentication
+from rest_framework.exceptions import MethodNotAllowed
 from rest_framework.pagination import PageNumberPagination
 
 from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession, Order
+from cinema.mixins import PermissionMixin
+
```
Dimosphen1: Do not make a blank line in case imports belong to similar import group according to PEP-8

Dimosphen1 (changes requested):

cinema/mixins.py
Marinel444: If I remove the mixin, it turns out that I need to use it in every place     def get_permissions(self):
        if self.action in ["retrieve", "update", "partial_update", "destroy"]:
            raise Http404
        if self.action == "create":
            return [IsAuthenticated]
        return super().get_permissions()  since IsAdminOrIfAuthenticatedReadOnlydoes not cover all permission

cinema/permissions.py
viktoria-rybenchuk: not fixed

cinema/permissions.py
viktoria-rybenchuk: not fixed

cinema/mixins.py
viktoria-rybenchuk: using `IsAdminOrIfAuthenticatedReadOnly ` permission is enough you don't need to redefine this logic in mixin

viktoria-rybenchuk (changes requested):

cinema/mixins.py
Marinel444: I did everything as you said but it doesn’t work for me, I don’t know how to do it without a mixin and with only one IsAdminOrIfAuthenticatedReadOnly

cinema/views.py
```diff
@@ -21,24 +23,49 @@
...
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
 
 
-class MovieViewSet(viewsets.ModelViewSet):
+class MovieViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.CreateModelMixin,
```
Dimosphen1, line 63: Place mixins on a first place in the inheritance in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #445, opened 2024-03-18
Sol

Dobryk12:


cinema/views.py
```diff
@@ -75,8 +101,18 @@ def get_serializer_class(self):
 
         return MovieSerializer
 
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
 
-class MovieSessionViewSet(viewsets.ModelViewSet):
```
juliastetsko, line 79: in this case you can inherit only `viewsets.ModelViewSet`

cinema/views.py
```diff
@@ -136,3 +178,12 @@ def get_serializer_class(self):
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+
+    def get_permissions(self):
+        if self.action in "create":
```
juliastetsko: it would be better `if self.action == "create"`

juliastetsko (changes requested):

cinema/views.py
```diff
@@ -136,3 +173,12 @@ def get_serializer_class(self):
...
         serializer.save(user=self.request.user)
+
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+
+    def get_permissions(self):
+        if self.action == "create":
+            return (IsAuthenticated(),)
+        else:
+            return super().get_permissions()
```
Serhii-Leonenko: This method is redundant in this case, use IsAuthenticated in the permission_classes instead of it

Serhii-Leonenko (changes requested):

cinema/views.py
```diff
@@ -1,10 +1,13 @@
...
+from rest_framework import viewsets, mixins
+from rest_framework.authentication import TokenAuthentication
 from rest_framework.pagination import PageNumberPagination
+from rest_framework.permissions import IsAuthenticated
 
 from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession, Order
+from cinema.permisions import IsAdminOrIfAuthenticatedReadOnly
 
 from cinema.serializers import (
     GenreSerializer,
```
Serhii-Leonenko: fix imports

Serhii-Leonenko (approved):

PR #462, opened 2024-03-20
Solution

MaxOperchuk:


PythonZem (approved):
gj

cinema/permissions.py
```diff
@@ -0,0 +1,25 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
+        )
```
MNDonut: request.user is always there, the only question is if he is authenticated.
Also what is the purpose of bool()? All your checks return boolean, thus you don't need to convert it to bool

cinema/views.py
```diff
@@ -118,12 +134,16 @@ class OrderPagination(PageNumberPagination):
...
+class OrderViewSet(mixins.CreateModelMixin,
+                   mixins.ListModelMixin,
+                   viewsets.GenericViewSet):
+
     queryset = Order.objects.prefetch_related(
         "tickets__movie_session__movie", "tickets__movie_session__cinema_hall"
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    permission_classes = (IsAuthenticatedOrReadOnly, )
```
MNDonut: Check the requirements. We didn't mention that we should give read only permission

MNDonut (changes requested):

Serhii-Leonenko (approved):

PR #463, opened 2024-03-20
solution

PythonZem:


AntonSkyrda (approved):

cinema/views.py
```diff
@@ -19,26 +28,52 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
MNDonut, line 42: One tip. Read about REST_FRAMEWORK constant
https://www.django-rest-framework.org/#example
It'll help you to make default authentication class, permissions and so on so you don't need to copy-paste this code everywhere. And it case you need to override somewhere it, you'll be able to do it

cinema/views.py
```diff
@@ -118,12 +157,16 @@ class OrderPagination(PageNumberPagination):
     max_page_size = 100
 
 
-class OrderViewSet(viewsets.ModelViewSet):
+class OrderViewSet(
+    viewsets.GenericViewSet, mixins.ListModelMixin, mixins.CreateModelMixin
```
MNDonut, line 161: separate line for each import

MNDonut (approved):

PR #465, opened 2024-03-21
Solution

JohnnyWalker010:


cinema/views.py
```diff
@@ -21,24 +23,45 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
```
Nick-r2c, line 24: Я не впевнений що потрібно було змінювати Model на Generic

Nick-r2c (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,13 @@
...
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
```
MNDonut: no need to check `if request.user`, cuz request user always is here, the only question is if he is authenticated or not
When you fix it, there won't be a need to use `bool` contructor

cinema/views.py
MNDonut: @Nick-r2c it's fine because ModelViewSet implemented all CRUD operations, but in checklist we have another requirement

user/serializers.py
```diff
@@ -1 +1,30 @@
...
+class UserSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = get_user_model()
+        fields = (
+            "id",
+            "username",
+            "email",
+            "password",
+            "is_staff",
+            "is_staff"
```
MNDonut: one is_staff is redundant

user/serializers.py
```diff
@@ -1 +1,30 @@
...
+
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password", None)
+        user = super().update(instance, validated_data)
+
+        if password:
+            user.set_password(password)
+            user.save()
+
+        return user
```
MNDonut, line 30: Is there a need to override these methods if you inherit from ModelSerializer?
Try to not override and check if it works

MNDonut (changes requested):
Also I see not all tests are passed, check them out and fix

user/serializers.py
JohnnyWalker010: Tests started failing so I couldn't remove these overrides.

MNDonut: My bad, yeah, you are right

cinema_service/settings.py
```diff
@@ -11,6 +11,8 @@
 """
 
 from pathlib import Path
+import rest_framework.authentication
+from rest_framework.authentication import TokenAuthentication
```
MNDonut: you don't need to do these imports. Django parses
`"rest_framework.authentication.TokenAuthentication"`
and finds this import itself. Basically that is the purpose of these settings where you add import path instead of importing classes :)

user/serializers.py
```diff
@@ -1 +1,29 @@
-# write your code here
+from django.contrib.auth import get_user_model
+from rest_framework import serializers
+
+
+class UserSerializer(serializers.ModelSerializer):
+    class Meta:
```
MNDonut: 1 blank line

user/views.py
```diff
@@ -1 +1,24 @@
...
+from user.serializers import UserSerializer
+
+
+class CreateUserView(generics.CreateAPIView):
+    serializer_class = UserSerializer
+
+
+class CreateTokenView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+
```
MNDonut: Here should be added implemented serializer, check the checklist out. If data is not correct, some errors should be raised

cinema/permissions.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
```
MNDonut: empty line

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,51 @@
...
+class CreateTokenView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+    serializer_class = UserSerializer
+
+    def post(self, request, *args, **kwargs):
+        username = request.data.get("username")
+        password = request.data.get("password")
+
+        if username and password:
```
Y-Havryliv: the `ObtainAuthToken` already handles that

Y-Havryliv (changes requested):

Serhii-Leonenko (approved):

PR #466, opened 2024-03-21
Solution

Nick-r2c:


cinema/views.py
```diff
@@ -124,6 +160,13 @@ class OrderViewSet(viewsets.ModelViewSet):
     )
     serializer_class = OrderSerializer
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (OrderPermission,)
```
mdubyna: u can use here built-in permission class (IsAuthenticated) also u can use generic and mixins to chose need actions, just like this 
class OrderViewSet(
    mixins.CreateModelMixin,
    mixins.ListModelMixin,
    viewsets.GenericViewSet
):

cinema/views.py
```diff
@@ -24,21 +30,49 @@
 class GenreViewSet(viewsets.ModelViewSet):
```
mdubyna: u can use generic here 
class GenreViewSet(
    mixins.CreateModelMixin,
    mixins.ListModelMixin,
    viewsets.GenericViewSet
):

cinema/views.py
```diff
@@ -24,21 +30,49 @@
...
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+
+    def get_queryset(self):
+        if self.action in ["list", "create"]:
+            return self.queryset
+        return None
 
 
 class ActorViewSet(viewsets.ModelViewSet):
```
mdubyna: also generic

cinema/views.py
```diff
@@ -24,21 +30,49 @@
...
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
+
+    def get_permissions(self):
+        if self.action in ["list", "create"]:
+            return (IsAdminOrIfAuthenticatedReadOnly(),)
+        raise NotFound
 
 
 class CinemaHallViewSet(viewsets.ModelViewSet):
```
mdubyna: also generic

mdubyna (changes requested):
use generic view and mixins for your actions

cinema/permissions.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class IsAdminOrIfAuthenticatedReadOnly(permissions.BasePermission):
+
+    def has_permission(self, request, view):
+        if (request.method in permissions.SAFE_METHODS
+                and request.user.is_authenticated):
+            return True
+
+        return request.user.is_authenticated and request.user.is_staff
```
MNDonut, line 11: It is not required to split to conditions here
You can make it as 1 if statement(because each check returns boolean) but that's good as well

cinema_service/settings.py
```diff
@@ -137,3 +137,9 @@
 # https://docs.djangoproject.com/en/4.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+REST_FRAMEWORK = {
+    "DEFAULT_AUTHENTICATION_CLASSES": [
+        "rest_framework.authentication.TokenAuthentication",
+    ]
+}
```
MNDonut, line 145: You can also add here DEFAULT_PERMISSION_CLASSES and set it to the most common permission. For viewset that don't need it, you can override permission_classes

MNDonut (approved):

PR #476, opened 2024-04-06
added login, register and me

MAFIOZNIK1999:


cinema/views.py
```diff
@@ -1,10 +1,14 @@
 from datetime import datetime
 
 from django.db.models import F, Count
-from rest_framework import viewsets
+from rest_framework import viewsets, mixins
+from rest_framework.authentication import TokenAuthentication
 from rest_framework.pagination import PageNumberPagination
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.viewsets import GenericViewSet
```
LiudmylaKulinchenko: You already have viewsets imported, u can use viewsets.GenericViewSet now

cinema/views.py
```diff
@@ -81,12 +102,15 @@ class MovieSessionViewSet(viewsets.ModelViewSet):
...
         .select_related("movie", "cinema_hall")
         .annotate(
-            tickets_available=F("cinema_hall__rows")
-            * F("cinema_hall__seats_in_row")
-            - Count("tickets")
+            tickets_available=F(
+                "cinema_hall__rows") * F(
+                "cinema_hall__seats_in_row") - Count(
+                "tickets")
         )
```
LiudmylaKulinchenko: You have bad formatting here
Parentheses should start and end with a text or have a newline before a text both at the start and the end

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good! ✨

PR #508, opened 2024-05-28
task: auth/token auth solution

panicua:


panicua:
I don't understand it (only on github):
```
Test Error
1s
Error: We are currently unable to download the log. Please try again later.
```
perhaps it's because now GPT servers are down + I cannot load screenshots for some reason even on github now (says wrong format but format is PNG, it's correct).

panicua:
The reason of the problem where chrome addon "ModHeader" that were changing authorization token and so it didn't let me see the log error of test on gitHub. :(

fsocie7y (approved):
seems good to me ^_^

PR #518, opened 2024-05-30
Solution 1

semermaks:


cinema/views.py
```diff
@@ -1,11 +1,13 @@
...
 from django.db.models import F, Count
-from rest_framework import viewsets
+from rest_framework.authentication import TokenAuthentication
 from rest_framework.pagination import PageNumberPagination
+from rest_framework import viewsets, mixins
+from rest_framework.permissions import IsAuthenticated
 
 from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession, Order
-
+from .permissions import IsAdminOrIfAuthenticatedReadOnly
```
MNDonut: if you use ".", then this import should be at the top of group

cinema/views.py
```diff
@@ -21,24 +23,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
+    authentication_classes = (TokenAuthentication,)
```
MNDonut: this is a good approach, but if you use the same auth_classes or permissions_classes almost everywhere, you can set up default classes in REST_FRAMEWORK constant. Here are docs: https://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme

user/views.py
```diff
@@ -1 +1,24 @@
...
+
+from user.serializers import UserSerializer
+
+
+class CreateUserView(generics.CreateAPIView):
+    serializer_class = UserSerializer
+
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
```
MNDonut: is it required to override rendered_classes if you override it by the same default renderer class?

MNDonut (changes requested):

fsocie7y (approved):
good job 👍

PR #519, opened 2024-05-30
Solution

SerhiiMar:


cinema/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS
+            and request.user
+            and request.user.is_authenticated
+        ) or bool(
+            request.user
+            and request.user.is_staff
+        )
```
MNDonut, line 14: 2 suggestions:
1. No need to check request.user because request always has user
2. No need to convert everything to boolean using bool(). All your conditions return boolean, thus the result will be always boolean

cinema/views.py
```diff
@@ -21,24 +26,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
MNDonut, line 37: this is a good approach, but if you use the same auth_classes or permissions_classes almost everywhere, you can set up default classes in REST_FRAMEWORK constant. Here are docs: https://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme

user/views.py
```diff
@@ -1 +1,23 @@
...
+
+from user.serializers import UserSerializer
+
+
+class CreateUserView(generics.CreateAPIView):
+    serializer_class = UserSerializer
+
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
```
MNDonut, line 14: is it required to override rendered_classes if you override it by the same default renderer class?

MNDonut (changes requested):

PR #520, opened 2024-05-30
Solution

Nyxan:


cinema/models.py
```diff
@@ -64,7 +64,8 @@ def __str__(self):
 class Order(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     user = models.ForeignKey(
-        settings.AUTH_USER_MODEL, on_delete=models.CASCADE
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE
```
MNDonut: add related_name

cinema/views.py
```diff
@@ -19,26 +21,46 @@
...
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.ListModelMixin, mixins.CreateModelMixin, viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminAllOrIsAuthenticatedOrReadOnly,)
```
MNDonut: this is a good approach, but if you use the same auth_classes or permissions_classes almost everywhere, you can set up default classes in REST_FRAMEWORK constant. Here are docs:
https://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy
https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme

user/permissions.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class IsAdminAllOrIsAuthenticatedOrReadOnly(BasePermission):
+
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS
+            and request.user
+            and request.user.is_authenticated
+        ) or (request.user and request.user.is_staff)
```
MNDonut: 2 suggestions:
1. No need to check request.user because request always has user
2. No need to convert everything to boolean using bool(). All your conditions return boolean, thus the result will be always boolean

user/serializers.py
```diff
@@ -1 +1,25 @@
-# write your code here
+from django.contrib.auth import get_user_model
+from rest_framework import serializers
```
MNDonut, line 2: fix imports

user/views.py
```diff
@@ -1 +1,24 @@
...
+
+from user.serializers import UserSerializer
+
+
+class CreateUserView(generics.CreateAPIView):
+    serializer_class = UserSerializer
+
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
```
MNDonut: is it required to override rendered_classes if you override it by the same default renderer class?

MNDonut (changes requested):

fsocie7y (approved):
well done ^_^

PR #522, opened 2024-06-01
Solution_for_py_token_authentication

GibbsVlad:


cinema/views.py
```diff
@@ -19,26 +27,52 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.CreateModelMixin,
+    mixins.ListModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
vsmutok: this is a good approach, but if you use the same auth_classes or permissions_classes almost everywhere, you can set up default classes in REST_FRAMEWORK constant. Here are docs:
https://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy
https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #533, opened 2024-06-10
Solution

anrozhna:


pavlopro (changes requested):
Why did you remove django toolbar?

anrozhna:
> pavlopro

because it's a good practice to remove the debug toolbar before release

pavlopro (approved):
Good Job!

PR #541, opened 2024-06-19
Solution

AlexRozv:


user/views.py
```diff
@@ -1 +1,25 @@
...
+from rest_framework.authtoken.views import ObtainAuthToken
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.settings import api_settings
+
+from user.serializers import UserSerializer
+
+
+class CreateUserView(generics.CreateAPIView):
+    serializer_class = UserSerializer
+    permission_classes = ()
```
LiudmylaKulinchenko: ?

LiudmylaKulinchenko (changes requested):

user/views.py
LiudmylaKulinchenko: I mean why is it blank 😁

AlexRozv: fixed and changed to AllowAny, it's needed because I used the custom permission class as default in settings, so we need to allow anyone to register

LiudmylaKulinchenko (approved):
Good job 🌼

PR #542, opened 2024-06-19
Develop

Anton-Konyk:


cinema/views.py
```diff
@@ -118,18 +140,20 @@ class OrderPagination(PageNumberPagination):
...
     pagination_class = OrderPagination
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated,)
 
     def get_queryset(self):
         return Order.objects.filter(user=self.request.user)
 
     def get_serializer_class(self):
-        if self.action == "list":
+        if self.request.method == "GET" and "pk" not in self.kwargs:
```
EdAlekseiev, line 156: Why? It looks like self.action == "list". Rollback these changes

EdAlekseiev (changes requested):

cinema/views.py
Anton-Konyk: Because self.action doesn't work with generics.ListCreateAPIView.  Action is attribute for object viewsets.ModelViewSet only.

Anton-Konyk: I think I can delete this condition ("pk" not in self.kwargs) because this class doesn't use retrieve.

EdAlekseiev (approved):

PR #548, opened 2024-07-06
solution

AlisaSkl:


cinema_service/settings.py
```diff
@@ -116,6 +116,15 @@
...
 
+# REST_FRAMEWORK = {
+#     "DEFAULT_PERMISSION_CLASSES": [
+#         "cinema.permissions.IsAdminOrIfAuthenticatedReadOnly"
+#     ],
+#     "DEFAULT_AUTHENTICATION_CLASSES": [
+#         "rest_framework.authentication.TokenAuthentication"
+#     ],
+# }
+
```
vsmutok: use it to avoid duplication

vsmutok (changes requested):

cinema_service/settings.py
AlisaSkl: I added default permission classes and removed duplication, but now I have some errors and I do not understand how to fix them

pavlopro (changes requested):
Tests Failed

viktoria-rybenchuk (approved):
GJ!

PR #562, opened 2024-07-26
solution

AndrewwSSS:


user/serializers.py
```diff
@@ -1 +1,36 @@
...
+
+    def create(self, validated_data):
+        return get_user_model().objects.create_user(**validated_data)
+
+    def update(self, instance, validated_data):
+        password = validated_data.pop("password", None)
+        if password:
+            instance.set_password(password)
+        super().update(instance, validated_data)
+        return instance
```
Arsen-hrynevych, line 36: You should call super().update() before setting the password to make sure the remaining validated data is processed correctly.

Arsen-hrynevych (approved):
Good job! I really appreciate your effort and attention to detail. Thank you for your hard work

PR #578, opened 2024-07-31
Solution

AntonKorinchuk:


cinema/views.py
```diff
@@ -1,10 +1,13 @@
 from datetime import datetime
 
 from django.db.models import F, Count
-from rest_framework import viewsets
+from rest_framework import viewsets, mixins
+from rest_framework.authentication import TokenAuthentication
 from rest_framework.pagination import PageNumberPagination
+from rest_framework.permissions import IsAuthenticated
 
 from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession, Order
```
Arsen-hrynevych, line 9: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

Arsen-hrynevych (approved):
Good job! I really appreciate your effort and attention to detail. Thank you for your hard work.

PR #580, opened 2024-07-31
solution

DimaYankovskiy:


cinema/views.py
```diff
@@ -1,11 +1,13 @@
...
 from django.db.models import F, Count
-from rest_framework import viewsets
+from rest_framework import viewsets, mixins
+from rest_framework.authentication import TokenAuthentication
 from rest_framework.pagination import PageNumberPagination
+from rest_framework.permissions import IsAuthenticated
 
-from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession, Order
-
+from cinema.models import Genre, Actor, Movie, CinemaHall, MovieSession, Order
```
Arsen-hrynevych, line 9: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

cinema_service/settings.py
```diff
@@ -137,3 +137,9 @@
 # https://docs.djangoproject.com/en/4.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+REST_FRAMEWORK = {
+    "DEFAULT_PERMISSION_CLASSES": [
+        "rest_framework.permissions.AllowAny",
+    ]
+}
```
Arsen-hrynevych, line 145: It's better to separate the logic into token and readonly rather than just using AllowAny and that's it.

user/urls.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from django.urls import path
+
+from user.views import CreateUserView, CreateTokenView, ManageUserViewSet
```
Arsen-hrynevych, line 3: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

Arsen-hrynevych (approved):
Good job! I really appreciate your effort and attention to detail. Thank you for your hard work.

PR #581, opened 2024-07-31
Solution

b3v3kt0r:


Arsen-hrynevych (approved):
Overall good, but please change the code format of classes in cinema.views.py.

PR #583, opened 2024-07-31
Solution

YevaYe:


cinema_service/settings.py
```diff
@@ -137,3 +137,9 @@
 # https://docs.djangoproject.com/en/4.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+REST_FRAMEWORK = {
+    "DEFAULT_PERMISSION_CLASSES": [
+        "rest_framework.permissions.AllowAny",
+    ]
+}
```
Arsen-hrynevych, line 145: It's better to separate the logic into token and readonly rather than just using AllowAny and that's it.

Arsen-hrynevych (approved):
Good job! I really appreciate your effort . Please review the check list of the task.

PR #584, opened 2024-08-01
'Solution'

Irina17191:


cinema/views.py
```diff
@@ -118,12 +155,19 @@ class OrderPagination(PageNumberPagination):
     max_page_size = 100
 
 
-class OrderViewSet(viewsets.ModelViewSet):
+class OrderViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    GenericViewSet,
+):
```
Arsen-hrynevych, line 162: It's correct.

[clarify]: it unnecessary to use mixins.ListModelMixin and mixins.CreateModelMixin separately when they can be combined into a single class. The ModelViewSet class from Django REST framework (DRF) provides both the list and create functionalities, among others. If you only need list and create actions, it's cleaner and more straightforward to use ModelViewSet and limit the HTTP methods accordingly.

Arsen-hrynevych (approved):
Thanks.

PR #585, opened 2024-08-01
solution

NikitaNikolaitsev:


Arsen-hrynevych (changes requested):
Overall good, but you need to fix the quotation issues, which you can see in the checks tab.

DevSpaciX (approved):

PR #588, opened 2024-08-04
"Solution"

MaglorFeanorson:


cinema_service/settings.py
```diff
@@ -137,3 +137,10 @@
 # https://docs.djangoproject.com/en/4.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+
+REST_FRAMEWORK = {
+    "DEFAULT_PERMISSION_CLASSES": [
+        "rest_framework.permissions.AllowAny",
+    ]
+}
```
Arsen-hrynevych: It would be better if you separate it into Token and Auth instead of using AllowAny and allow any.

Arsen-hrynevych (changes requested):
Perfect, but fix minor issues. Thanks.

Arsen-hrynevych (approved):
Thanks

PR #590, opened 2024-08-05
'Solution'

Vitalii2906:


cinema/views.py
```diff
@@ -21,118 +22,161 @@
...
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def update(self, request, pk=None):
+        try:
+            movie_session = MovieSession.objects.get(pk=pk)
+            serializer = MovieSessionSerializer(
+                movie_session, data=request.data
+            )
+            if serializer.is_valid():
```
vasylhnatiuk: use serializer.is_valid(raise_exception=True)

vasylhnatiuk (changes requested):

cinema/views.py
```diff
@@ -21,118 +22,161 @@
...
+            movie_session = MovieSession.objects.get(pk=pk)
+            serializer = MovieSessionSerializer(
+                movie_session, data=request.data
+            )
+            if serializer.is_valid(raise_exception=True):
+                serializer.save()
+                return Response(serializer.data, status=status.HTTP_200_OK)
+            return Response(
+                serializer.errors, status=status.HTTP_400_BAD_REQUEST
+            )
```
EdAlekseiev: it you use `is_valid` with raise_exception=True you shouldn't use if. Fix everywhere, please
```suggestion
            serializer.is_valid(raise_exception=True):
            serializer.save()
            return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #594, opened 2024-08-07
sol2

IvanVahanov:


cinema/urls.py
```diff
@@ -1,13 +1,8 @@
...
 from rest_framework import routers
-
 from cinema.views import (
-    GenreViewSet,
-    ActorViewSet,
-    CinemaHallViewSet,
-    MovieViewSet,
-    MovieSessionViewSet,
-    OrderViewSet,
+    GenreViewSet, ActorViewSet, CinemaHallViewSet, MovieViewSet,
```
maxgerman, line 4: style

cinema/views.py
```diff
@@ -21,24 +28,48 @@
...
-class CinemaHallViewSet(viewsets.ModelViewSet):
+class CinemaHallViewSet(
+    mixins.CreateModelMixin,
+    mixins.ListModelMixin,
+    GenericViewSet,
+):
     queryset = CinemaHall.objects.all()
     serializer_class = CinemaHallSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
maxgerman, line 61: these authentication and permission classes could be made default in the settings instead of placing them in each view

maxgerman (changes requested):
Good job!
You shouldn't commit directly to your master/main branch - use the branch as in the task.
See some comments

PR #598, opened 2024-08-19
Solution

IraMladanovych02:


pavlopro (changes requested):
Tests failed
PR for check issue

cinema/views.py
```diff
@@ -21,22 +30,33 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class ListCreateModelsViewMixin(
```
Dimosphen1: The new class is not a mixin since it inherits from `GenericViewSet` as well. I'll suggest removing this class and inherit all views from necessary mixins explicitly to provide better customization in the future

Dimosphen1 (changes requested):

cinema_service/settings.py
```diff
@@ -12,6 +12,8 @@
 
 from pathlib import Path
 
+from cinema.permissions import IsAdminOrIfAuthenticatedReadOnly
```
pavlopro: Why did you import this here?

cinema/views.py
```diff
@@ -21,24 +34,30 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(GenericViewSet, ListModelMixin, CreateModelMixin):
```
pavlopro: Please inherit statring from mixin classes

cinema/views.py
```diff
@@ -21,24 +34,30 @@
...
-class MovieViewSet(viewsets.ModelViewSet):
+class MovieViewSet(
+    GenericViewSet,
+    ListModelMixin,
+    CreateModelMixin,
+    RetrieveModelMixin,
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = (TokenAuthentication,)
```
pavlopro: No need to have this

cinema/views.py
```diff
@@ -76,7 +95,7 @@ def get_serializer_class(self):
         return MovieSerializer
 
 
-class MovieSessionViewSet(viewsets.ModelViewSet):
+class MovieSessionViewSet(ModelViewSet):
```
pavlopro: use viewsets.ModelViewSet and remove unused import

pavlopro (changes requested):

LLkaia (approved):

PR #601, opened 2024-08-21
solution

daikj88:


cinema/views.py
```diff
@@ -21,24 +31,49 @@
...
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
```
pavlopro: These authentication and permission classes could be made default in the settings instead of placing them in each view

cinema/views.py
```diff
@@ -21,24 +31,49 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
```
pavlopro, line 34: Place mixins on the first place

pavlopro (changes requested):

EdAlekseiev (approved):

PR #602, opened 2024-08-23
Solution

Sergunshot:


cinema/views.py
```diff
@@ -19,26 +21,52 @@
...
-class ActorViewSet(viewsets.ModelViewSet):
+class ActorViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Actor.objects.all()
     serializer_class = ActorSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
EdAlekseiev: These authentication and permission should be default and placed in settings to avoid code duplication

EdAlekseiev (changes requested):

cinema/views.py
Sergunshot: Done

EdAlekseiev (approved):

PR #607, opened 2024-09-11
Implement solution

dimak20:


cinema/permissions.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS and request.user
+            and request.user.is_authenticated
+        ) or (
+            request.user and request.user.is_staff
+        )
```

cinema/views.py
```diff
@@ -21,24 +23,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = [TokenAuthentication, ]
+    permission_classes = [IsAdminOrIfAuthenticatedReadOnly, ]
```

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,28 @@
...
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+    authentication_classes = []
+    permission_classes = []
+
+
+class ManageUserView(generics.RetrieveUpdateAPIView):
+    serializer_class = UserSerializer
+    authentication_classes = [TokenAuthentication, ]
```

MNDonut (changes requested):

user/views.py
dimak20: My carelessness

MNDonut (approved):

PR #608, opened 2024-09-14
solution

Fedot0v:


cinema/permissions.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            (request.method in SAFE_METHODS
+             and request.user
+             and request.user.is_authenticated
+             )
```
LLkaia:
```suggestion
            (
                 request.method in SAFE_METHODS
                 and request.user
                 and request.user.is_authenticated
             )
```

cinema/views.py
```diff
@@ -136,3 +178,8 @@ def get_serializer_class(self):
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+    def get_permissions(self):
+        if self.action == "create":
+            return (IsAuthenticated(),)
+        return super().get_permissions()
```
LLkaia, line 179: permission_classes = [IsAuthenticated,]

Use this instead

LLkaia (commented):
Instead of modifying each view, configure auth in settings.py
Find info here: https://www.django-rest-framework.org/api-guide/authentication/

LLkaia (changes requested):
Do requested changes

cinema/views.py
```diff
@@ -21,24 +23,45 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+):
```
LLkaia, line 30: What the reason of this changes in each view?

LLkaia (changes requested):

cinema/views.py
Fedot0v: because we only need some part of CRUD operations

LLkaia (approved):

PR #612, opened 2024-09-26
Solution

Depaton:


cinema/views.py
```diff
@@ -21,24 +24,49 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
```
Dimosphen1: It is a good practice to place mixins in the first position in the inheritance. Please, apply this change in all places

cinema/views.py
```diff
@@ -81,12 +109,14 @@ class MovieSessionViewSet(viewsets.ModelViewSet):
         MovieSession.objects.all()
         .select_related("movie", "cinema_hall")
         .annotate(
-            tickets_available=F("cinema_hall__rows")
-            * F("cinema_hall__seats_in_row")
-            - Count("tickets")
+            tickets_available=F("cinema_hall__rows"
+                                ) * F("cinema_hall__seats_in_row"
+                                      ) - Count("tickets")
```
Dimosphen1: There is no need to place closing parenthesis no a new line, please return to the previous version so that there would not be a case that there is inconsistency in the presence of arguments on the same line with opening and closing parenthesis

cinema/views.py
```diff
@@ -21,24 +24,49 @@
...
-class MovieViewSet(viewsets.ModelViewSet):
+class MovieViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = [TokenAuthentication, ]
```
Dimosphen1: It would be better to allocate tuple rather than list for storing classes for `authentication_classes` and `permission_classes`. Please, apply this change in all places

user/views.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from rest_framework import generics
+from user.serializers import UserSerializer
+from rest_framework.authtoken.views import ObtainAuthToken
+from rest_framework.settings import api_settings
+from rest_framework.authentication import TokenAuthentication
+from rest_framework.permissions import IsAuthenticated
```
Dimosphen1: Please, order and group imports according to PEP-8 in all places

Dimosphen1 (changes requested):

cinema/views.py
```diff
@@ -17,28 +27,53 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication, )
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
```
pavlopro: these authentication and permission classes could be made default in the settings instead of placing them in each view

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #614, opened 2024-09-29
implement py-token-authentication

olenazaritska:


cinema/urls.py
```diff
@@ -18,6 +18,8 @@
 router.register("movie_sessions", MovieSessionViewSet)
 router.register("orders", OrderViewSet)
 
-urlpatterns = [path("", include(router.urls))]
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
Dimosphen1, line 23: In case `urlpatterns` consist of only router urls, you could assign `router.urls` to `urlpatterns` directly. Follow this approach in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #615, opened 2024-09-29
solution

BastovOleksandr:


cinema/permissions.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS
+            and request.user
```
MasakDirt: you can create field and reuse it instead of all time getting it from request

MasakDirt (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,12 @@
...
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        user = request.user
+        auth_constraints = all(
+            (request.method in SAFE_METHODS, user, user.is_authenticated)
+        )
+        staff_constraints = all((user, user.is_staff))
```
MNDonut: there is always user in request, thus no need to check it

cinema/views.py
```diff
@@ -21,24 +24,43 @@
...
 
 
-class CinemaHallViewSet(viewsets.ModelViewSet):
+class CinemaHallViewSet(
+    mixins.ListModelMixin, mixins.CreateModelMixin, viewsets.GenericViewSet
+):
     queryset = CinemaHall.objects.all()
     serializer_class = CinemaHallSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```

MNDonut (changes requested):

MNDonut (approved):

PR #617, opened 2024-10-02
Solution

yaroslavsysoiev:


cinema/views.py
```diff
@@ -21,32 +24,54 @@
...
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.CreateModelMixin,
+    mixins.ListModelMixin,
+    GenericViewSet,
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
```
pavlopro: Instead of modifying each view, configure auth in settings.py
Find info here: https://www.django-rest-framework.org/api-guide/authentication/

pavlopro (changes requested):

EdAlekseiev (approved):

PR #619, opened 2024-10-02
Task Solution

achalaidiuk:


DSahalatyi (approved):

cinema_service/settings.py
```diff
@@ -116,6 +116,12 @@
 
 AUTH_USER_MODEL = "user.User"
 
+REST_FRAMEWORK = {
+    "DEFAULT_AUTHENTICATION_CLASSES": [
+        "rest_framework.authentication.TokenAuthentication",
+    ]
+}
```
MNDonut: you can set default permissions here as well and override it in views where it should be different

user/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+
+    def has_permission(self, request, view):
+        return bool(
+            (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
```
MNDonut: user is always in request

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,25 @@
...
+    permission_classes = ()
+
+
+class CreateTokenView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+
+
+class ManageUserView(generics.RetrieveUpdateAPIView):
+    serializer_class = UserSerializer
+    authentication_classes = (TokenAuthentication,)
```

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from rest_framework import generics
+from rest_framework.authentication import TokenAuthentication
```
MNDonut: you didn't remove import ;)

MNDonut (changes requested):

cinema/views.py
```diff
@@ -21,24 +23,45 @@
...
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
MNDonut: you set permission class but still override by the base permission

MNDonut (changes requested):

MNDonut (approved):

PR #620, opened 2024-10-02
Solution

DSahalatyi:


panichqa (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+
+    def has_permission(self, request, view):
+        return bool(
+            (request.user and request.user.is_staff)
+            or (
+                request.user
+                and request.user.is_authenticated
```

user/views.py
```diff
@@ -1 +1,26 @@
...
+
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+    permission_classes = (AllowAny,)
+
+
+class ManageUserView(generics.RetrieveUpdateAPIView):
+    serializer_class = UserSerializer
+    authentication_classes = (TokenAuthentication,)
```

MNDonut (changes requested):

MNDonut (approved):

PR #621, opened 2024-10-02
First commit

panichqa:


cinema/permissions.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminAllORAuthenticatedORReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        user = request.user
+        auth_constraints = all(
+            (request.method in SAFE_METHODS, user, user.is_authenticated)
```
vladislav-tsybuliak1: It is better to put each constraint on separate line.

user/urls.py
```diff
@@ -1 +1,11 @@
-# write your code here
+from django.urls import path
+
+from user.views import CreateUserView, LoginUserView, ManageUserView
+
```
vladislav-tsybuliak1, line 4: Add 1 blank line here. Between import statements and the rest of the code - 2 blank lines.

vladislav-tsybuliak1 (approved):
Great job! Keep moving on!

cinema/permissions.py
MNDonut: there is always user in request, thus no need to check it

cinema/views.py
```diff
@@ -21,24 +24,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminAllORAuthenticatedORReadOnly,)
```

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,25 @@
...
+    permission_classes = ()
+
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+
+
+class ManageUserView(generics.RetrieveUpdateAPIView):
+    serializer_class = UserSerializer
+    authentication_classes = (TokenAuthentication,)
```

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from rest_framework import generics
+from rest_framework.authentication import TokenAuthentication
```

MNDonut (changes requested):

user/views.py
panichqa: :(

MNDonut (approved):

PR #623, opened 2024-10-02
Develop

vladislav-tsybuliak1:


cinema/views.py
```diff
@@ -21,24 +23,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
OleksiukStepan: You can define DEFAULT_AUTHENTICATION_CLASSES and DEFAULT_PERMISSION_CLASSES in settings to avoid duplication

OleksiukStepan (approved):

cinema/views.py
vladislav-tsybuliak1: Thanks a lot for the advice! This looks much better now!

user/views.py
```diff
@@ -1 +1,26 @@
...
+    permission_classes = ()
+
+
+class LoginUserView(ObtainAuthToken):
+    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
+
+
+class ManageUserView(generics.RetrieveUpdateAPIView):
+    serializer_class = UserSerializer
+    authentication_classes = (TokenAuthentication,)
```

MNDonut (changes requested):

MNDonut (approved):

PR #625, opened 2024-10-02
first_try

yashkunn:


ValeriiaKyr (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,15 @@
...
+    The request is authenticated as a user, or is a read-only request.
+    """
+
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS and request.user
+            and request.user.is_authenticated
+        ) or (
+            request.user and request.user.is_staff
+        )
```

cinema/views.py
```diff
@@ -21,24 +24,49 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```

MNDonut (changes requested):

cinema/permissions.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    """
+    The request is authenticated as a user, or is a read-only request.
+    """
+
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS and request.user
```
MNDonut: you still check user

cinema/views.py
```diff
@@ -21,24 +23,45 @@
...
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
```
MNDonut: why do you override auth classes by default auth class when you set default one in settings?

MNDonut: fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #626, opened 2024-10-03
task done

Sergi0bbb:


anton-bilous (approved):

cinema/permissions.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS
+            and request.user
+            and request.user.is_authenticated
+        ) or (
+            request.user and request.user.is_staff
```
MNDonut: user is always in request

cinema/views.py
```diff
@@ -21,24 +24,48 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreListCreateViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```

MNDonut (changes requested):

user/views.py
```diff
@@ -1 +1,27 @@
-# write your code here
+from rest_framework import generics
+from rest_framework.authentication import TokenAuthentication
```

MNDonut (changes requested):

MNDonut (approved):

PR #628, opened 2024-10-03
Solution --Test--

dartomOOv:


cinema_service/settings.py
```diff
@@ -125,7 +125,7 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```
achalaidiuk, line 128: You need to add default authentication and permission classes

cinema_service/permissions.py
```diff
@@ -0,0 +1,13 @@
...
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            request.method in SAFE_METHODS
+            and request.user
+            and request.user.is_authenticated
+        ) or (
+            request.user and request.user.is_staff
+        )
```
achalaidiuk: user is always in request

achalaidiuk (changes requested):

user/views.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from rest_framework import generics
+from rest_framework.authentication import TokenAuthentication
```

MNDonut (changes requested):

MNDonut (approved):

PR #636, opened 2024-10-04
'Solution'

Poznanskyi:


olena4reunited (approved):

user/views.py
```diff
@@ -1 +1,25 @@
-# write your code here
+from django.contrib.auth import get_user_model
+from rest_framework import generics
+from rest_framework.authentication import TokenAuthentication
```

MNDonut (changes requested):

cinema/permissions.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    """
+    If is authenticated as an admin - read/write;
+    if is authenticated as a user - read only.
+    """
+
+    def has_permission(self, request: Request, view: View) -> bool:
+        return bool(
```
MNDonut: do you need bool here taking into account that all operators return boolean?

MNDonut (changes requested):

MNDonut (approved):

PR #639, opened 2024-10-05
Solution

AndriiHamasa:


cinema/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+    def has_permission(self, request, view):
+        view_name = view.__class__.__name__
+        view_action = view.action
+
+        if view_name in [
+            "ActorViewSet",
+            "GenreViewSet",
+            "CinemaHallViewSet",
+        ] and view_action in ["retrieve", "destroy", "update"]:
+            raise NotFound("Object not found.")
```
Maksym-Turenko: It's better to use `PermissionDenied` instead of `NotFound`

cinema/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+            and request.user
+            and request.user.is_staff
+        )
+
+
+class ReadAndCreateIfAuthenticated(BasePermission):
+
+    def has_permission(self, request, view):
+        if view.action in ["delete", "update", "retrieve", "destroy"]:
+            raise NotFound("Object not found.")
```
Maksym-Turenko: `PermissionDenied` instead of `NotFound`

Maksym-Turenko (changes requested):

cinema/permissions.py
AndriiHamasa: `PermissionDenisd`  has code 403, but in tests we need 404

MNDonut: are you sure we had to implement it?

cinema/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+                view_action in ("list", "retrieve")
+                and request.user
+                and request.user.is_authenticated
+            ) or (request.user and request.user.is_staff)
+
+        return bool(
+            view_action == "list"
+            and request.user
+            and request.user.is_authenticated
+        ) or (request.user and request.user.is_staff)
```
MNDonut: request always contains user

cinema/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+            request.method in ("GET", "POST")
+            and request.user
+            and request.user.is_staff
+        )
+
+
+class ReadAndCreateIfAuthenticated(BasePermission):
+
+    def has_permission(self, request, view):
+        if view.action in ["delete", "update", "retrieve", "destroy"]:
```
MNDonut: are you sure you need to implement these classes?

cinema/views.py
```diff
@@ -24,21 +30,29 @@
 class GenreViewSet(viewsets.ModelViewSet):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
MNDonut: instead of duplicating code, you can set DEFAULT_PERMISSION_CLASSES and DEFAULT_AUTHENTICATION_CLASSES in REST_FRAMEWORK constant in settings.py and override these attributes where needed

MNDonut (changes requested):

cinema/permissions.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework.exceptions import NotFound, MethodNotAllowed
```
MNDonut: do you use these imports?

MNDonut (changes requested):

cinema/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    """The request is authenticated as an admin - read/write,
+    if as a user - read only request"""
+
+    def has_permission(self, request, view):
+
+        return bool(
+            request.method in SAFE_METHODS
+            and request.user.is_authenticated
+        ) or request.user.is_staff
```
MNDonut: "in" operator returns boolean, is_authenticated as well, so what is the purpose to convert it to bool?

MNDonut (changes requested):

MNDonut (approved):

PR #644, opened 2024-10-06
Solution

raychw:


cinema/views.py
```diff
@@ -1,10 +1,21 @@
...
-from rest_framework import viewsets
+from rest_framework import viewsets, mixins
 from rest_framework.pagination import PageNumberPagination
-
-from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession, Order
+from rest_framework.authentication import TokenAuthentication
+from rest_framework.permissions import IsAuthenticated
+
+from cinema_service.permissions import IsAdminOrIfAuthenticatedReadOnly
+
```
Dimosphen1: There is no need to add a blank line between imports that belong to the same import group according to PEP-8, remove it, please

cinema/views.py
```diff
@@ -21,24 +32,49 @@
 )
 
 
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
```
Dimosphen1: It is a better approach to place mixins on a first position in inheritance here and in the other classes as well

cinema_service/settings.py
Dimosphen1, line 1: You could add in settings.py to use `IsAdminOrIfAuthenticatedReadOnly` and `TokenAuthentication` as default permission and authentication classes to not add them in all views for the `cinema` app

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #645, opened 2024-10-06
Solution

dpiuro:


Bogdan-Rylov (approved):

cinema/views.py
```diff
@@ -14,45 +23,64 @@
...
-class GenreViewSet(viewsets.ModelViewSet):
+class GenreViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
MNDonut: instead of duplicating code, you can set DEFAULT_PERMISSION_CLASSES and DEFAULT_AUTHENTICATION_CLASSES in REST_FRAMEWORK constant in settings.py and override these attributes where needed

user/permissions.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import permissions
+
+
+class IsAdminOrIfAuthenticatedReadOnly(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if request.user and request.user.is_staff:
```
MNDonut: there is always user in request

MNDonut (changes requested):

user/views.py
```diff
@@ -1,7 +1,8 @@
 from rest_framework import generics
 from rest_framework.authentication import TokenAuthentication
```
MNDonut: do you need this import taking into account the it's a default auth class?

MNDonut (changes requested):

MNDonut (approved):

PR #662, opened 2024-11-06
solution

oleg-gj:


cinema_service/settings.py
```diff
@@ -12,6 +12,8 @@
 
 from pathlib import Path
 
+from user.permissions import IsAdminOrIfAuthenticatedReadOnly
```
pavlopro, line 15: imported but not used

cinema/views.py
```diff
@@ -21,24 +25,43 @@
...
+class MovieViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    viewsets.GenericViewSet,
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
pavlopro: Instead of modifying each view, configure authentication_classes and permission_classes in settings.py

pavlopro (changes requested):

vasylhnatiuk (approved):

PR #681, opened 2024-12-07
Solution

VladimirDolhyi:


cinema/views.py
```diff
@@ -21,24 +25,49 @@
...
+class MovieViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    GenericViewSet,
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
Dimosphen1: Please, indicate a default permission class in the `settings.py`, as well as default  authentication class to avoid code duplication. You could explicitly redefine these entities in the corresponding views in case needed

Dimosphen1 (changes requested):

vsmutok (approved):
Good Job!

PR #682, opened 2024-12-07
solution

popeye88:


cinema/views.py
```diff
@@ -1,44 +1,74 @@
...
+class MovieViewSet(
+    ListModelMixin,
+    CreateModelMixin,
+    RetrieveModelMixin,
+    GenericViewSet,
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
vsmutok: Please, indicate a default permission class in the `settings.py`, as well as default authentication class to avoid code duplication. You could explicitly redefine these entities in the corresponding views in case needed

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #695, opened 2024-12-13
Solution-py-token-authentication

skyfoxwork:


cinema/views.py
```diff
@@ -21,24 +24,49 @@
...
+class MovieViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin
+):
     queryset = Movie.objects.prefetch_related("genres", "actors")
     serializer_class = MovieSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
pohgen, line 69: you can write it in settings and not dublicate it across all views

pohgen (approved):
except for one flaw, you've made a good solution

PR #698, opened 2024-12-14
solution

sofiiasavkova:


cinema/views.py
```diff
@@ -17,66 +21,91 @@
...
 
-class MovieSessionViewSet(viewsets.ModelViewSet):
+class MovieSessionViewSet(
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+    GenericViewSet,
+):
```
Atikiho: You can leave MovieSessionViewSet as ModelViewSet as it already implements all these actions

Atikiho (approved):

PR #709, opened 2024-12-16
Solution

yulia-pl:


cinema/views.py
```diff
@@ -21,32 +24,70 @@
...
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
 
+    def get_permissions(self):
+        if self.action == "create":
+            self.permission_classes = [IsAdminUser]
+        else:
+            self.permission_classes = [IsAuthenticated]
+        return super().get_permissions()
```
alauda6, line 38: As I understand task, you already have implemented IsAdminORIfAuthenticatedReadOnly. That's the same what you wrote in def get_permissions. So like python dev - probably should just wrote permission_classes = (IsAdminOrIfAuthenticatedReadOnly,) and no need to write for every ViewSet +7 code lines, just add one attribute.

alauda6 (approved):
Your code worked, but next time should write it in more simple way

PR #716, opened 2024-12-18
Solution

FursykIvan:


AnyoneClown (changes requested):
Your tests didn't pass because of dependency conflict between `django-debug-toolbar`, `djangorestframework`, and `Django`

Use older version of `djangorestframework` that was provided in your repo

Nattalli (approved):

PR #747, opened 2025-02-12
Solution

Dogober:


cinema/permissions.py
```diff
@@ -0,0 +1,23 @@
...
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (
+                request.method
+                in (
+                    "GET",
+                    "POST",
+                )
+                and view.basename == "order"
```
ruslanyemets, line 18: You could simplify has_permission method and not use this class for OrderViewSet

cinema/views.py
```diff
@@ -118,7 +141,11 @@ class OrderPagination(PageNumberPagination):
     max_page_size = 100
 
 
-class OrderViewSet(viewsets.ModelViewSet):
+class OrderViewSet(
+    viewsets.GenericViewSet, mixins.ListModelMixin, mixins.CreateModelMixin
+):
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
ruslanyemets, line 148: You could use here built_in permission class

ruslanyemets (approved):
Good job!

PR #752, opened 2025-02-13
Solution

bodiakof:


VovaGld (commented):
create solution for task

yelizaveta-zh (changes requested):
implement solution

PR #759, opened 2025-02-14
Solution

Professor-Douel:


cinema/permissions.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class IsAdminOrAuthenticatedReadOnly(BasePermission):
+
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS and request.user.is_authenticated:
```
Nattalli:
```suggestion
        if 
        request.method in SAFE_METHODS and request.user and request.user.is_authenticated:
```

cinema/permissions.py
```diff
@@ -0,0 +1,9 @@
...
+            return True
+        return request.user.is_authenticated and request.user.is_staff
```
Nattalli: you can union this both conditions with `or` station

Nattalli (changes requested):

Nattalli (approved):

PR #760, opened 2025-02-14
Create task Token authentication

ihorhalyskiy:


cinema/permissions.py
```diff
@@ -0,0 +1,23 @@
...
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (
+                request.method
+                in (
+                    "GET",
+                    "POST",
+                )
+                and view.basename == "order"
```
Nattalli: why?

Nattalli: that is bad solution, better use  `permission_classes = None` in the view

Nattalli (changes requested):

Nattalli (approved):

PR #776, opened 2025-03-08
Develop

saywin:


cinema/views.py
```diff
@@ -87,6 +115,8 @@ class MovieSessionViewSet(viewsets.ModelViewSet):
         )
     )
     serializer_class = MovieSessionSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
```
vsmutok, line 119: this is a good approach, but if you use the same auth_classes or permissions_classes almost everywhere, you can set up default classes in REST_FRAMEWORK constant. Here are docs: [https://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme](https://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy%E2%80%A8https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme)

vsmutok (changes requested):

PR #799, opened 2025-04-17
Solution

lyaskov:


lyaskov:
@mateacademy-ai-mentor
The variables app_name = "cinema" and app_name = "user" are already present at the end of the cinema/urls.py and user/urls.py files, respectively. Please double-check again.

PR #805, opened 2025-04-22
Anton Zhukov - Solution

windings-lab:


windings-lab:
@mateacademy-ai-mentor 
About raise NotFound(). In task it says: "all tests should pass. user/tests & cinema/tests". So based on that I made raising NotFound(), because otherwise mentor's tests are not gonna work

PR #812, opened 2025-04-28
completed

BGBroth:


BGBroth:
@mateacademy-ai-mentor

1. if i dont change allowany permission to another one than creating a new account will be imposible cause unauthorized user will not have permissions to create one
2. IsAdminOrIfAuthenticatedReadOnly applied to all viewsets by default cause i added it to settings.py
3. cinema/urls.py defines app_name = "cinema"

PR #813, opened 2025-04-28
Solution

iSevenpwnz:


cinema/views.py
```diff
@@ -21,24 +24,55 @@
...
+class IsAdminOrIfAuthenticatedReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return bool(
+            (
+                request.method in SAFE_METHODS
+                and request.user
+                and request.user.is_authenticated
+            )
+            or (request.user and request.user.is_staff)
+        )
```
Yurkiv313: Not a mistake, but it would be better to create a file named permission.py for custom permissions

cinema/views.py
```diff
@@ -21,24 +24,55 @@
...
+
+
+class GenreViewSet(
+    mixins.CreateModelMixin, mixins.ListModelMixin, viewsets.GenericViewSet
```
Yurkiv313, line 39: You can create a custom mixin that inherits from 
`mixins.CreateModelMixin`, `mixins.ListModelMixin`, and `viewsets.GenericViewSet`, 
and reuse this mixin across your next classes

cinema/views.py
```diff
@@ -21,24 +24,55 @@
...
+):
     queryset = Genre.objects.all()
     serializer_class = GenreSerializer
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly,)
 
 
-class ActorViewSet(viewsets.ModelViewSet):
+class ActorViewSet(
+    mixins.CreateModelMixin, mixins.ListModelMixin, viewsets.GenericViewSet
```
Yurkiv313, line 48: the same

Yurkiv313 (approved):
gj

PR #820, opened 2025-04-29
Solution

Baranotik15:


user/serializers.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from django.contrib.auth import get_user_model
+from rest_framework import serializers
+
+
+class UserSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = get_user_model()
```
Boyok1337: Define it one time at start and reuse it

user/serializers.py
```diff
@@ -1 +1,24 @@
...
+class UserSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = get_user_model()
+        fields = ("id", "username", "email", "password", "is_staff")
+        read_only_fields = ("id", "is_staff")
+        extra_kwargs = {"password": {"write_only": True, "min_length": 5}}
+
+    def create(self, validated_data):
+        return get_user_model().objects.create_user(**validated_data)
```
Boyok1337: The same

cinema/views.py
```diff
@@ -1,138 +1,147 @@
...
+        return Order.objects.filter(user=self.request.user)
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return OrderListSerializer
+
+        return OrderSerializer
+
+    def perform_create(self, serializer):
+        serializer.save(user=self.request.user)
```
Boyok1337: If u create viewsets.py, move this to viewsets.py.

cinema/views.py
```diff
@@ -1,138 +1,147 @@
...
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
+
+
+class OrderPagination(PageNumberPagination):
+    page_size = 10
+    max_page_size = 100
```
Boyok1337: This is not view, so move to other file

Boyok1337 (changes requested):

Boyok1337 (approved):
GJ!

PR #873, opened 2025-06-24
Solution

Hlibwork:


Hlibwork:
@mateacademy-ai-mentor 2, 3, 4 In principle, they do not apply to this task, they have already been implemented earlier and were well received. 1 I don't think is fair, since the code passes all the tests and is written according to the requirements.

PR #907, opened 2025-09-04
Token Authentication API

farzamgt:


cinema/views.py
```diff
@@ -19,62 +23,98 @@
...
+        if self.action in ["retrieve", "update", "partial_update", "destroy"]:
+            from rest_framework.exceptions import NotFound
+            raise NotFound()
+        return super().get_object()
 
 
 class ActorViewSet(viewsets.ModelViewSet):
     queryset = Actor.objects.all()
     serializer_class = ActorSerializer
+    permission_classes = [IsAdminOrIfAuthenticatedReadOnly]
```
DanilHamster, line 44: ❗️The methods retrieve, update, partial_update, and destroy are overridden to return 404, which appears to be a test bypass rather than an architectural decision. It would be cleaner to restrict allowed actions using http_method_names or appropriate mixins.

cinema/views.py
```diff
@@ -19,62 +23,98 @@
...
+    def partial_update(self, request, *args, **kwargs):
+        from rest_framework.response import Response
+        return Response(status=405)
+
+    def destroy(self, request, *args, **kwargs):
+        from rest_framework.response import Response
+        return Response(status=405)
+
 
 class MovieSessionViewSet(viewsets.ModelViewSet):
```
DanilHamster, line 119: ❗️The methods retrieve, update, partial_update, and destroy are overridden to return 404, which appears to be a test bypass rather than an architectural decision. It would be cleaner to restrict allowed actions using http_method_names or appropriate mixins.

DanilHamster (approved):
Approve but check comments and fix your code

PR #918, opened 2025-09-07
Solution

viannik:


VadymKyrylko (approved):
LGFM

viannik:
@mateacademy-ai-mentor IsAdminOrIfAuthenticatedReadOnly is set as REST_FRAMEWORK.DEFAULT_PERMISSION_CLASSES. Therefore, it applies to everything. And that is why IsAuthenticated is set in orders, because it does not cause an error. Everything has been done as required, please confirm the successful completion of the task.

PR #922, opened 2025-09-07
Solution

Arch0998:


Orixara (approved):

Arch0998:
@mateacademy-ai-mentor 
In the first review, ReadCreateOnlyViewSet.http_method_names was marked as an important recommendation, not a critical blocker, while here it’s listed as critical.

There’s some duplication regarding OrderViewSet.get_permissions() — the core issue is the same, but the wording and details differ.

This second review also adds new critical points (TokenAuthentication in settings, extra manage/ route) which were not mentioned in the first review.

If I follow the recommendations exactly as described here, some tests will fail, because certain suggestions conflict with the requirements as tested.

PR #931, opened 2025-09-08
Solution

Plovuezz:


Vskesha (approved):
good solution. All the requirements were implemented. The code is clear. Great work!

Plovuezz:
@mateacademy-ai-mentor permission is called IsAdminOrIfAuthenticatedReadOnly so only authetificated user can do safe methods such as get, what are you talking about? Please approve

PR #932, opened 2025-09-08
Solution

DenisPrudius:


Salmon-eat (approved):

DenisPrudius:
@mateacademy-ai-mentor 
Thanks for the feedback!
According to the project guideline (checklist item 3.3), authenticated users must be able to create orders, while anonymous users are restricted.
For that reason I intentionally used IsAuthenticated in OrderViewSet instead of the global IsAdminOrIfAuthenticatedReadOnly.
The custom permission still applies to all other cinema endpoints as required.
This keeps the behavior consistent with the tests: only logged-in users can create orders, and admins retain full access.
Since all tests (cinema + user) are passing, this implementation matches the requirements.

DenisPrudius (commented):
@mateacademy-ai-mentor 
Thanks for the feedback!
According to the project guideline (checklist item 3.3), authenticated users must be able to create orders, while anonymous users are restricted.
For that reason I intentionally used IsAuthenticated in OrderViewSet instead of the global IsAdminOrIfAuthenticatedReadOnly.
The custom permission still applies to all other cinema endpoints as required.
This keeps the behavior consistent with the tests: only logged-in users can create orders, and admins retain full access.
Since all tests (cinema + user) are passing, this implementation matches the requirements.

PR #940, opened 2025-09-10
Fix solution

Sir-Churchill:


Sir-Churchill:
2. If we another one we broke test so i need to use IsAuthenticated or test will fail
3. I have spaces in a end of file

PR #941, opened 2025-09-12
Solution

MaksymBus:


cinema/views.py
```diff
@@ -21,24 +24,49 @@
...
     serializer_class = ActorSerializer
+    authentication_classes = (TokenAuthentication, )
+    permission_classes = (IsAdminOrIfAuthenticatedReadOnly, )
 
 
-class CinemaHallViewSet(viewsets.ModelViewSet):
+class CinemaHallViewSet(
+    viewsets.GenericViewSet,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin
```
Dimosphen1, line 52: It is a common practice to place mixins before base classes. Change the inheritance order in the future, so that mixins will be in a first place.

Dimosphen1 (approved):
Looks good to me, well done!

PR #1003, opened 2026-01-03
Solution

dmytro-malyk-dm:


vakt159 (approved):
Looks good, but you could've add REST_FRAMEWORK = {
 "DEFAULT_PERMISSION_CLASSES":  .... IsAdminOrIfAuthenticatedReadOnly } to your settings.py and you wouldn't need to do that mixin class and permission_class in your views.

PR #1020, opened 2026-01-05
Solution

Akcolotle:


user/serializers.py
```diff
@@ -1 +1,44 @@
-# write your code here
+from django.contrib.auth import get_user_model
+from rest_framework import serializers
+
+User = get_user_model()
+
+
+class UserRegisterSerializer(serializers.ModelSerializer):
```
leemonch1k-space: In this case you can use one serializer for all actions

leemonch1k-space (approved):
Great job!

PR #1124, opened 2026-06-19
Solution

idubina:


idubina:
@mateacademy-ai-mentor Thanks for the review. I refactored the cinema viewsets to use the shared custom permission instead of manually overriding get_permissions in multiple places.

Regarding anonymous access for SAFE_METHODS: I tested this against the provided test suite, and the tests fail when SAFE_METHODS are allowed for unauthenticated users. Because of that, I kept SAFE_METHODS available only for authenticated users, while non-safe methods remain restricted to staff/admin users.

PR #1131, opened 2026-06-24
Token Authentication

Shongi:


Shongi:
@mateacademy-ai-mentor
"deletion via the API must be prohibited even for administrators, but your current MovieSessionViewSet still exposes the destroy action."

This literally contradicts your own test for admin:

def test_delete_movie_session(self):
        movie_session = sample_movie_session()

        url = detail_url(movie_session.id)
        response = self.client.delete(url)

        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)

